PR #159, opened 2026-06-01
feat(workflows): branch-based Wait execution with nested waits

Lorchie:


lightningpixel:
1. label logic duplicated and inconsistent between WaitNode.tsx and WorkflowPanel.tsx

WaitNode.tsx:


const label = waitState === 'done' ? 'Retry' : waitState === 'error' ? 'Retry' : 'Continue'
WorkflowPanel.tsx:


const label = waitState === 'done' || waitState === 'error' ? 'Retry' : 'Continue'
Same output by coincidence, but the redundant ternary in WaitNode will diverge silently next time someone edits it. Unify these — and while you're at it, canContinue is also copy-pasted verbatim between the two components. Both pieces of logic should live in one place (a shared helper or exposed from the store).

2. _ctx module-level mutable state breaks under Vite HMR


const _ctx = { current: null as RunContext | null }
A hot reload during development resets _ctx.current to null while an async run is still in flight. The next call to continueRun(id) finds no context and either throws or fails silently. Add a guard with a clear message:


if (!_ctx.current) {
  console.warn('continueRun: no active run context — was the module hot-reloaded mid-run?')
  return
}
Nits
preflight.ts hardcodes type strings despite the new registry


if ((node.type === 'extensionNode' || node.type === 'outputNode') && ...)
Now that nodeBehaviors.ts exists, this should use the registry predicates instead of string literals — otherwise adding a new node type means updating two places.

reachesSceneOutput only traverses passthrough nodes forward

The function won't walk through intermediate extensionNodes — so extensionNode → extensionNode → outputNode returns false. If that's intentional (i.e. only direct or passthrough-bridged paths count), add a comment explaining why. If not, the traversal condition needs to be broadened.

lightningpixel (approved):

PR #160, opened 2026-06-01
feat(settings): accessibility section

Lorchie:


lightningpixel:
1. TextSize is defined twice

App.tsx redefines a type that is already used inline in appStore.ts:


// App.tsx — should not be here
type TextSize = 'small' | 'medium' | 'large'
Export the type from appStore.ts and import it in App.tsx. That also removes the need for the as TextSize cast on the next line, which is only there to work around the duplication.

2. Missing type="button" on SegmentedControl buttons

In ui/index.tsx, the <button> elements have no explicit type. Browsers default to type="submit" inside a form, which would trigger an unintended submission. Add type="button" to each one.

3. SegmentedControl has no ARIA attributes — ironic for an Accessibility section

The component has no semantic markup for assistive technologies:

The container should have role="group" and accept an aria-label prop (e.g. "Text size")
Each active button should have aria-pressed={active}
4. Import alignment in SettingsPage.tsx

The extra spaces added to align the existing imports are not consistent with the rest of the file. Either apply it everywhere or remove it — as-is it just adds noise to the diff.

lightningpixel (approved):

PR #173, opened 2026-06-08
Add transform gizmo (move/rotate/scale)

### LINKED ISSUE #161: FEATURE 1 - Interactive transform gizmo
## FEATURE 1 — Interactive transform gizmo (move / rotate / scale in scene)

  **As a user viewing a generated mesh,** I want to grab a gizmo and move,
  rotate, or scale the mesh directly in the 3D scene, so I can position it
  visually without typing numeric values or leaving the app.

  Complements the numeric `POST /optimize/transform` ticket — this is the
  interactive counterpart (previously marked "Tier 3 / out of scope").

  ### Scope
  - Gizmo attached to the loaded mesh when it is **selected** (selection state
    already exists in `Viewer3D.tsx`).
  - Three modes: **Translate / Rotate / Scale**, switchable via toolbar buttons
    and keyboard (`W` / `E` / `R`).
  - Gizmo dragging disables OrbitControls (drei handles this automatically when
    `OrbitControls` has `makeDefault`).
  - Optional snapping (grid / 15° / 0.1) while holding a modifier key.
  - "Apply" bakes the gizmo transform into the mesh so it persists to export;
    "Reset" reverts to the auto-centered pose.

  ### Implementation
  - **Viewer** `src/areas/generate/components/Viewer3D.tsx`:
    - Add `<TransformControls>` from `@react-three/drei`, attached to the mesh
      `<primitive object={scene}>` via a ref (lift a ref out of `SceneMeshModel`).
    - Render only when `selected`. Mode driven by new state `gizmoMode:
      'translate' | 'rotate' | 'scale' | null`.
    - Reconcile with the existing auto-center `useEffect` (Viewer3D.tsx:201) so a
      user transform isn't overwritten on re-render (only auto-center on first
      load / model change).
  - **Toolbar** `ViewerToolbar.tsx`: add a Move/Rotate/Scale button group
    (mirrors the existing `MODES` button pattern), plus `gizmoMode` props.
  - **Keyboard**: extend the existing keydown handler (Viewer3D.tsx:396) with
    W/E/R to switch mode and `Esc` to exit gizmo (skip when an input is focused).
  - **Bake/persist (Apply)**: read the mesh's world matrix and send it to the
    backend to bake into the GLB. Reuse / extend `POST /optimize/transform`
    (`api/routers/optimize.py`) by accepting a raw 4×4 `matrix`, applying it at
    scene level with `Scene.apply_transform()` (preserves materials/UV), then
    `updateCurrentJob` + `pushMeshUrl` to reload the viewer on the result.

  ### Acceptance criteria
  - [ ] Selecting the mesh shows a gizmo; deselecting hides it.
  - [ ] Translate / rotate / scale all work and update the mesh live.
  - [ ] W/E/R switch modes; toolbar buttons reflect the active mode.
  - [ ] OrbitControls is disabled while dragging the gizmo, re-enabled after.
  - [ ] "Apply" bakes the transform into the GLB; re-loading keeps the new pose.
  - [ ] "Reset" restores the auto-centered position.
  - [ ] No interference with viewMode switching, screenshot, or Delete-to-remove.

  **Estimate:** M

iammojogo-sudo:
Closes #161

Adds a gizmo to move/rotate/scale the mesh in the viewport. Shows when the
mesh is selected, W/E/R to switch modes. Apply bakes it into the GLB so it
sticks on export, Reset reverts.

New /optimize/transform endpoint on the backend handles the bake.

Tested locally, works.


lightningpixel:
You’ll need to resolve the conflicts on the PR

Edit: 
And I added the new toolbar on the Generate page.
So instead of putting the buttons in the bar on the left side, you should use the second toolbar at the top.
I already added the transform buttons there. You’ll just need to adapt it

lightningpixel:
The pushed branch isn't in a working state, so I have to request changes.

  ## Blocking issues

  1. **The branch doesn't run.** `Viewer3D.tsx` uses `gizmoMode`, `setGizmoMode`, `applying`, `resetToken`, and `meshObject` but never declares them (no `useState`), and `TransformControls` is used without being imported from `@react-three/drei`. Running the app immediately
  throws `Uncaught ReferenceError: setGizmoMode is not defined`, and `npm run build` fails type-checking. It looks like part of your wiring commit was lost before pushing — could you push the complete version you actually tested?

  2. **`api/routers/optimize.py` contains the entire module twice** (paste/merge accident — `router = APIRouter` appears at lines 25 and 361, every endpoint is defined twice). The second `router` shadows the first, and since it's the old copy without `/transform`, the new
  endpoint is never registered: every "Apply" call will 404. Please reduce the file to a single copy that includes `/transform`.

  3. **Gizmo wiring is incomplete.** `MeshModelProps` declares `onObject` but `MeshModel` never calls it, so `meshObject` would stay `null` and the gizmo would never render. `SceneMeshModel` reads `autoCenter` and `resetToken` but those props are never added to its signature nor
  passed down.

  4. **Out-of-scope changes** — please remove these from the PR:
     - The README "Sponsors" section and the X/Twitter links. Sponsor recognition is a maintainer decision; adding yourself there isn't something we can merge from a feature PR.
     - The version bumps in `package.json` and `api/main.py` — `dev` has already moved to 0.3.6 independently, so this will conflict anyway.

  ## Smaller things to fix in the same pass

  - The `{stem}_xf.glb` output name means re-applying produces `..._xf_xf.glb`, and applying twice from the same base overwrites the file while keeping the same URL — `useGLTF`'s cache will then show the stale mesh. A unique suffix (we already use `uuid` elsewhere) would avoid
  both problems.
  - `/transform` reimplements its path check inline — please reuse `_resolve_input_path` like `/mesh` and `/smooth` do.
  - `handleApplyTransform` has no `catch`, so a backend failure is silent for the user.
  - The baked matrix includes the auto-centering offset applied on load, so "Apply" permanently moves the file's origin even for a pure rotation. If that's intentional ("bake what you see"), a comment explaining it would help.
  - The W/E/R/Escape handler only guards against `HTMLInputElement`/`HTMLTextAreaElement` — contenteditable elements would still trigger mode switches.
  - `useApi.ts` uses tab indentation against the file's spaces, and `importMesh` was re-indented without changes — this will fail `npm run lint`.

  Happy to re-review once the complete, building version is pushed. Thanks again for the contribution !

iammojogo-sudo:
> The pushed branch isn't in a working state, so I have to request changes.
> ## Blocking issues
> 
>     1. **The branch doesn't run.** `Viewer3D.tsx` uses `gizmoMode`, `setGizmoMode`, `applying`, `resetToken`, and `meshObject` but never declares them (no `useState`), and `TransformControls` is used without being imported from `@react-three/drei`. Running the app immediately
>        throws `Uncaught ReferenceError: setGizmoMode is not defined`, and `npm run build` fails type-checking. It looks like part of your wiring commit was lost before pushing — could you push the complete version you actually tested?
> 
>     2. **`api/routers/optimize.py` contains the entire module twice** (paste/merge accident — `router = APIRouter` appears at lines 25 and 361, every endpoint is defined twice). The second `router` shadows the first, and since it's the old copy without `/transform`, the new
>        endpoint is never registered: every "Apply" call will 404. Please reduce the file to a single copy that includes `/transform`.
> 
>     3. **Gizmo wiring is incomplete.** `MeshModelProps` declares `onObject` but `MeshModel` never calls it, so `meshObject` would stay `null` and the gizmo would never render. `SceneMeshModel` reads `autoCenter` and `resetToken` but those props are never added to its signature nor
>        passed down.
> 
>     4. **Out-of-scope changes** — please remove these from the PR:
>        
>        * The README "Sponsors" section and the X/Twitter links. Sponsor recognition is a maintainer decision; adding yourself there isn't something we can merge from a feature PR.
>        * The version bumps in `package.json` and `api/main.py` — `dev` has already moved to 0.3.6 independently, so this will conflict anyway.
> 
> 
> ## Smaller things to fix in the same pass
> 
>     * The `{stem}_xf.glb` output name means re-applying produces `..._xf_xf.glb`, and applying twice from the same base overwrites the file while keeping the same URL — `useGLTF`'s cache will then show the stale mesh. A unique suffix (we already use `uuid` elsewhere) would avoid
>       both problems.
> 
>     * `/transform` reimplements its path check inline — please reuse `_resolve_input_path` like `/mesh` and `/smooth` do.
> 
>     * `handleApplyTransform` has no `catch`, so a backend failure is silent for the user.
> 
>     * The baked matrix includes the auto-centering offset applied on load, so "Apply" permanently moves the file's origin even for a pure rotation. If that's intentional ("bake what you see"), a comment explaining it would help.
> 
>     * The W/E/R/Escape handler only guards against `HTMLInputElement`/`HTMLTextAreaElement` — contenteditable elements would still trigger mode switches.
> 
>     * `useApi.ts` uses tab indentation against the file's spaces, and `importMesh` was re-indented without changes — this will fail `npm run lint`.
> 
> 
> Happy to re-review once the complete, building version is pushed. Thanks again for the contribution !

Not a worry, thank you and Ill work on this today until its ready. Sorry for the extra work. I hope to reduce that soon.

iammojogo-sudo (commented):
I noticed that drhepa's feature for hightlighting meshes wasnt in this pull. I just wanted to note that I pulled a few hours ago and the last submit was about 6 hours ago so its recent but I didnt see his code in there anymore. Hope that helps!

iammojogo-sudo:
Also for this one (And sorry for the questions, I just want to make sure of things). So this section is the concern: 
```<<<<<<< main
  // Gizmo hotkeys: W/E/R switch tool, Esc exits. Ignored while typing.
  useEffect(() => {
    const handler = (e: KeyboardEvent) => {
      const el = document.activeElement as HTMLElement | null
      if (el && (el instanceof HTMLInputElement || el instanceof HTMLTextAreaElement || el.isContentEditable)) return
      if (e.key === 'Escape') { setGizmoMode((m) => (m ? null : m)); return }
      if (!hasModel || !meshSelected) return
      const k = e.key.toLowerCase()
      if (k === 'w') setGizmoMode('translate')
      else if (k === 'e') setGizmoMode('rotate')
      else if (k === 'r') setGizmoMode('scale')
    }
    window.addEventListener('keydown', handler)
    return () => window.removeEventListener('keydown', handler)
  }, [hasModel, meshSelected])
=======
  useEffect(() => {
    if (openPanel !== 'library' || libraryLoaded || libraryLoading) return
    void loadLibraryEntries()
  }, [openPanel, libraryLoaded, libraryLoading])
>>>>>>> dev
```

It appears here that should we keep both? Would we need to actually merge these instead of removing either? The libraryLoaded seems like it has to do with another feature, and so my additions were specific to the requested keystrokes that were needed. SO, I just wnat to know, should I accept both OR were there changes/conflicts before that we has made that already resolves the removal of the current section? 

Thank you I appreciate your patience with me and time! God Bless you :)

iammojogo-sudo (commented):
Roger that :)

PR #181, opened 2026-06-10
Sun light UI scrollbar setting parameter persistence

iammojogo-sudo:
Persist 3D viewer lighting settings (Sun / Fill) across sessions.

The lighting panel sliders reset to defaults on every page load.
This saves them to localStorage on change and restores them on
startup, so they persist across generations, tab switches, and
app restarts until the user changes them or hits Reset.

One file changed: GeneratePage.tsx.   
image

### DIFF
--- a/src/areas/generate/GeneratePage.tsx
+++ b/src/areas/generate/GeneratePage.tsx
@@ -1,7 +1,7 @@
 import { useState, useRef, useCallback, useEffect } from 'react'
 import type { ReactNode } from 'react'
-import { useAppStore } from '@shared/stores/appStore'
-import type { GenerationJob } from '@shared/stores/appStore'
+import { useAppStore, DEFAULT_LIGHT_SETTINGS } from '@shared/stores/appStore'
+import type { GenerationJob, LightSettings } from '@shared/stores/appStore'
 import { useApi } from '@shared/hooks/useApi'
 import { ColorPicker } from '@shared/components/ui'
 import GenerationHUD from './components/GenerationHUD'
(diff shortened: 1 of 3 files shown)


Lorchie:
Thanks for the PR. A couple of points before this can be merged:

**1. Doesn't follow the existing persistence pattern**

The project already persists state through the zustand `persist` middleware (`src/shared/stores/appStore.ts:133`, key `modly-store`, with `partialize` for `generationOptions`, `uiScale`, etc.). This PR reimplements persistence by hand (manual `localStorage` get/set + `JSON.parse` + field-by-field validation + error handling) when zustand already handles all of that.

`lightSettings` should live in `appStore` and be added to its `partialize`. Otherwise we end up with two competing persistence mechanisms, which is harder to maintain.

**2. Diff noise / out-of-scope changes (~9 of ~13 hunks)**

Most of the diff is unrelated to persistence — it's Unicode character replacements in UI labels and comments:
- `Processing…` → `Processing...` (×4), `Importing…` → `Importing...`
- `(1–20)` → `(1-20)` (en-dash → hyphen)
- `—` → `-` in comments (×3)

This looks like an editor encoding corruption. It pollutes the review, contradicts the "one file changed: lighting persistence" description, and degrades the UI text (the typographic ellipsis `…` replaced by three dots).

Please revert these to the original characters so the PR only contains the `localStorage` + `useEffect` + `useState` init block.

Once persistence is moved into `appStore` and the encoding changes are reverted, this should be good to go.

iammojogo-sudo:
Will do :)

iammojogo-sudo (commented):
Changes made per request "persistence is moved into appStore and the encoding changes are reverted," If there are still issues, let me know :)

Lorchie:
Perfect !!! :)

lightningpixel (approved):

PR #193, opened 2026-06-20
patch/keep image inputs in order on multi-image nodes

iammojogo-sudo:
If an extension node has multiple image inputs (front/left/back/right
style), skipping a middle one used to shift everything after it up a
slot — e.g. wire front + left + right but skip back, and right would
end up getting treated as back.

This keeps each image tied to its actual handle, so a skipped slot just
stays empty instead of bumping the rest. Only touches workflowRunStore.ts.

Tested locally with a multi-image extension, all the skip combos work right.

### DIFF
--- a/src/areas/workflows/nodes/ExtensionNode.tsx
+++ b/src/areas/workflows/nodes/ExtensionNode.tsx
@@ -120,11 +120,8 @@ export default function ExtensionNode({ id, data, selected }: { id: string; data
   const { updateNodeData } = useReactFlow()
   const running = useWorkflowRunStore((s) => s.activeNodeId === id)
 
-  // Refs for handle alignment — support up to 2 inputs
-  const ioRowRef  = useRef<HTMLDivElement>(null)
-  const ioRow2Ref = useRef<HTMLDivElement>(null)
-  const [handleTop,  setHandleTop]  = useState('50%')
-  const [handle2Top, setHandle2Top] = useState('50%')
+  const handleRefs = useRef<HTMLDivElement[]>([])
+  const [handleTops, setHandleTops] = useState<string[]>([])
 
   const { modelExtensions, processExtensions } = useExtensionsStore()
   const allExtensions = buildAllWorkflowExtensions(modelExtensions, processExtensions)
@@ -138,15 +135,14 @@ export default function ExtensionNode({ id, data, selected }: { id: string; data
 
   // Align handles with their respective IO rows after mount
   useLayoutEffect(() => {
-    if (ioRowRef.current) {
-      const center = ioRowRef.current.offsetTop + ioRowRef.current.offsetHeight / 2
-      setHandleTop(`${center}px`)
-    }
-    if (ioRow2Ref.current) {
-      const center = ioRow2Ref.current.offsetTop + ioRow2Ref.current.offsetHeight / 2
-      setHandle2Top(`${center}px`)
-    }
-  }, [isMulti])
+    setHandleTops(handleRefs.current.map((ref) => {
+      if (ref) {
+        const center = ref.offsetTop + ref.offsetHeight / 2
+        return `${center}px`
+      }
+      return '50%'
+    }))
+  }, [isMulti, inputs?.length])
 
   const patchParam = useCallback((key: string, val: number | string) => {
     updateNodeData(id, { params: { ...data.params, [key]: val } })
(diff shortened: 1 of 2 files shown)


iammojogo-sudo:
No need to add before 0.4 :)

iammojogo-sudo:
NOTE: 
This, Im sure you know isn't vital. It's only required for likely some extensions that handle multiple inputs, which are only a few. Thats why it was updated after the other three.

lightningpixel:
### Code review

I dug into this one and the problem isn't the work itself — it's what the merge in `e5eac44` did to it.

`0f2b7ff` had a complete, coherent implementation: `extraImagePaths` declared, slot-aware routing off `edge.targetHandle`, and `processParams` actually passed to `runProcess`. Merging `dev` resolved three of those regions in dev's favour and left a fourth untouched, which gutted the feature but left fragments behind:

| Region | Resolved as | Result |
|---|---|---|
| Declaration block | dev | gains `nodeInputTexts`, **loses `extraImagePaths`** |
| Edge loop | dev | **loses the slot-aware routing entirely** |
| `runProcess` call | dev | gains `texts:`/`liveParams`, **stops passing `processParams`** |
| `processParams` construction | branch | dev had nothing here, no conflict, **survives orphaned** |

So the undefined identifiers below aren't typos — they're the scars of lines deleted elsewhere in the same merge. `ExtensionNode.tsx` took the same hit: the `handleRefs` refactor survives in the declarations and handles, but the multi-input `ioSubheader` reverted to dev's version that still uses `ioRowRef`/`ioRow2Ref`.

Concretely, on the current head:

1. The slot routing is gone. `inputTypes`/`inputPaths` are declared but never populated or read, and the loop below still keys on `src.outputType`, overwriting `nodeInputPath` per edge — so the last edge wins and skipping a middle slot still shifts everything.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/workflowRunStore.ts#L317-L331

2. `extraImagePaths` is referenced three times, declared nowhere. This sits in the generic process-extension branch, so it breaks every process node, not just multi-image ones.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/workflowRunStore.ts#L432-L440

3. `processParams` is built and discarded — the call still passes `liveParams`, so `extra_image_paths` never reaches the backend.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/workflowRunStore.ts#L441-L450

4. `ioRowRef`/`ioRow2Ref` are still used but no longer declared. Beyond the reference itself, the multi-input rows never register into `handleRefs`, so `handleTops` stays empty and every handle falls back to `top: '50%'` — alignment is lost precisely in the multi-input case.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/nodes/ExtensionNode.tsx#L206-L228

5. Only half the multi-input UI got generalized. Handles `.map()` over all N inputs, but the label rows still hardcode `inputs[0]` and `inputs[1]`, so the 4-image front/left/back/right case renders handles 2 and 3 with no row — despite the comment on L206 saying "one row per input".

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/nodes/ExtensionNode.tsx#L249-L260

To unblock: items 1-3 are literally code that existed in `0f2b7ff` and can be restored, keeping dev's genuine additions (`nodeInputTexts`, `liveParams`) — the two sides are compatible, they just touch the same lines. Only item 5 needs something new.

Worth noting `mergeable: MERGEABLE` on this PR only means git can apply the patch cleanly; the conflict was already resolved in `e5eac44`, so that green flag says nothing about correctness here.

One small thing: the description says "Only touches workflowRunStore.ts", but `ExtensionNode.tsx` is modified too.

iammojogo-sudo:
Oh wow yeah that def is likely not worth aything yet then at this point! Would you like for me to rework it with a more current version or even think about doing it over after the next update maybe? I dont want it to mess things up!

lightningpixel:
### Code review

Found 5 issues:

1. `extraImagePaths` is referenced but never declared, breaking every process-extension node run

`processParams.extra_image_paths = [nodeInputPath, ...extraImagePaths]` and `extraImagePaths.length > 0` run unconditionally in the non-model branch of `executeExtensionNode` — i.e. on every process-extension execution, not just multi-image ones. No `extraImagePaths` declaration exists anywhere in the file.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/workflowRunStore.ts#L432-L439

2. `ioRowRef`/`ioRow2Ref` are used as refs but were removed as declarations, breaking render of any multi-input node

The refactor replaced `ioRowRef`/`ioRow2Ref` with a single `handleRefs` array, but the multi-input branch of `ioSubheader` still does `ref={ioRowRef}` / `ref={ioRow2Ref}` — undeclared identifiers.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/nodes/ExtensionNode.tsx#L206-L225

3. The multi-image slot-ordering bug this PR claims to fix is still present

`inputPaths` is declared to hold one path per input slot but is never written to or read. The edge loop still does `nodeInputPath = src.filePath` for every image edge (last edge wins), with no indexing by `edge.targetHandle` the way the neighboring text-input case does on L329. Skipping a middle image slot still shifts the rest, unchanged from before the PR.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/workflowRunStore.ts#L316-L333

4. `processParams` (meant to carry `extra_image_paths`) is built and then discarded

`runProcess` is still called with `liveParams` as its params argument, not `processParams` — so even once `extraImagePaths` is declared, the extra image paths never reach the backend.

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/workflowRunStore.ts#L433-L450

5. `ioSubheader` still hardcodes only `inputs[0]`/`inputs[1]` while `handlesEl` now renders N handles

`handlesEl` was generalized to `inputs.map(...)` for any number of inputs, but the labeled header rows above it were not — only two rows exist. For an extension with 3+ image inputs, the extra handles render with no corresponding labeled row, despite the comment above stating "one row per input."

https://github.com/lightningpixel/modly/blob/e5eac44c0fd090372248e8e7e07db5649b9713af/src/areas/workflows/nodes/ExtensionNode.tsx#L206-L228

lightningpixel (approved):

PR #205, opened 2026-06-23
feat/add single-image preview node, fix text-only detection for multi-input models

iammojogo-sudo:
## Files changed
src/areas/workflows/nodes/ImagePreviewNode.tsx — new (created + our update)
src/areas/workflows/WorkflowsPage.tsx — registration (1 line add)
src/areas/workflows/preflight.ts — registration (1 line add)

## New: ImagePreviewNode
A general single-image preview node for workflows. Reads the output file
from disk via `window.electron.fs.readFileBase64()` and renders as a
`data:` URL — avoids CSP issues and API origin dependency. Has a
pass-through image output so it can sit inline between nodes.

## Fix: text-only model detection
A model is text-only only if **all** its declared inputs are `"text"`.
Previously the check only looked at the first input, which incorrectly
treated multi-input models like Flux/Qwen (`"inputs": ["text", "image"]`)
as text-only and sent a 1×1 placeholder instead of the real image.

## Additional
- Simplified multi-input routing to use `src.outputType` instead of slot-based matching
- Added `extraParams.text` alongside `extraParams.prompt` for generator compatibility

----------------------------------------------------
image

----------------------------------------------------
image

### DIFF
--- a/src/areas/workflows/WorkflowsPage.tsx
+++ b/src/areas/workflows/WorkflowsPage.tsx
@@ -29,6 +29,7 @@ import TextNode         from './nodes/TextNode'
 import AddToSceneNode   from './nodes/AddToSceneNode'
 import Load3DMeshNode   from './nodes/Load3DMeshNode'
 import PreviewImageNode from './nodes/PreviewImageNode'
+import ImagePreviewNode from './nodes/ImagePreviewNode'
 import WaitNode         from './nodes/WaitNode'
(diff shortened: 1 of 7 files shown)


iammojogo-sudo (commented):
Ok so there will be a ForEachNode addition to the preview, or will that be specifically a variable used for statement checking?

lightningpixel (approved):

iammojogo-sudo:
Rock on!

PR #209, opened 2026-06-27
play a short ding sound when generation or workflow run completes

iammojogo-sudo:
When running a generation (image-to-3D or a multi-node workflow), the app currently gives no audible feedback when it finishes. If you're in another tab or just looking away, there's no way to know it's done without checking back.

This PR adds a short, unobtrusive 880Hz "ding" sound (think microwave or oven beep) played once when:

A single image-to-3D generation finishes
A workflow run finishes (all branches complete)
The sound is generated on the fly using the Web Audio API — no audio files to bundle, no new dependencies, works in any modern browser/Electron environment. If audio isn't available for some reason, it silently does nothing.

Changes are minimal: a new utility file (src/shared/utils/sound.ts) and one line each in useGeneration.ts and workflowRunStore.ts.

### DIFF
--- a/electron/main/ipc-handlers.ts
+++ b/electron/main/ipc-handlers.ts
@@ -1,4 +1,4 @@
-import { ipcMain, BrowserWindow, dialog, app, shell } from 'electron'
+import { ipcMain, BrowserWindow, Notification, dialog, app, shell } from 'electron'
 import { buildSync } from 'esbuild'
 import { autoUpdater } from 'electron-updater'
 import { join } from 'path'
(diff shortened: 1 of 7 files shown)


lightningpixel:
Thanks for the idea, it’s really interesting !

But I was imagining something a bit different. Instead of adding a small sound, I was thinking more about a notification system, for example using native Windows notifications :)

iammojogo-sudo:
You are very welcome! Would you like for me to update this as well as its title so that it would be more visual notifications as well?

iammojogo-sudo (commented):
Whatever changes you make are ok with me

lightningpixel (approved):

iammojogo-sudo:
Thanks brother! I hope it wasnt difficult to make any modifications and I apologize if I hadn't been around much lately. Hoepfully the codes help a lot and in the future I might just stay fully AI if its easier lol.

PR #211, opened 2026-06-28
Fixed swapped icons on Import/Export buttons in GeneratePage.tsx

ivk1800:
Fixed swapped icons on Import/Export buttons in GeneratePage.tsx

Import button now shows upload icon (arrow pointing up/out of box)
Export button now shows download icon (arrow pointing down/into box)

The fix was needed because the icons were semantically reversed:

Import = bringing data in → should use download icon (arrow pointing down into box)
Export = sending data out → should use upload icon (arrow pointing up out of box)
The original code had them swapped - Import showed the upload icon, Export showed the download icon. This is a common UX mistake since "up = upload" and "down = download" is the standard convention.


Lorchie:
Of course thx you !!

lightningpixel (approved):

PR #229, opened 2026-07-18
Feat/light updates

iammojogo-sudo:
I have not made updates yet but I wanted to get the PR in there. I did update to teh 0.4.1 branch though.


iammojogo-sudo (commented):
Looks better and patch addition will be there so tahts good. I noticed other changes that were 'additions' so as long as they are good then lets do it :)

iammojogo-sudo:
Oh if you want me to reupdate with a 0.4.1 or 0.4.2 udpate for it (whichever one we are one right now), I can do that and if you want me to change/update any features I can do that before commits.

lightningpixel:
### Code review

Found 2 issues:

1. Gizmo mode silently re-activates on mesh→point-light→mesh reselection, regressing a bug fixed at commit 2b5ad32 (the reset effect's own comment: "Drop the active transform tool when nothing is selected, so it doesn't silently re-activate on the next selection"). The reset effect only clears `gizmoMode` when both `meshSelected` and `selectedPointLightId` are falsy, but neither `handleMeshSelect` nor the point-light `onSelect` handler ever clear `gizmoMode` themselves. So: select a mesh, press R (rotate), then select a point light (mesh deselects, effect doesn't fire because `selectedPointLightId` is now truthy), then click back onto the mesh — `gizmoMode` is still `'rotate'` and the Rotate tool (and its toolbar button) reactivates without the user pressing R again.

https://github.com/lightningpixel/modly/blob/8b80923720fa2bad96980fbed5b3246e2ba0b195/src/areas/generate/GeneratePage.tsx#L683-L697

2. Unrelated changes bundled into this feature PR — version bumps in `package.json`/`api/main.py`, a new `build-macos` CI job, and README edits (typo fix, new "Workflows" section, star-history embed URL/token change) — none of which relate to the PR's stated point-light feature. This repeats a pattern the repo owner already flagged on PR #173: "Out-of-scope changes — please remove these from the PR: ... The version bumps in `package.json` and `api/main.py` — `dev` has already moved to 0.3.6 independently, so this will conflict anyway." `main` has since moved to 0.4.2 and already has a `build-macos` job, so these bundled changes are now stale/redundant and will conflict.

https://github.com/lightningpixel/modly/blob/8b80923720fa2bad96980fbed5b3246e2ba0b195/package.json#L1-L5
https://github.com/lightningpixel/modly/blob/8b80923720fa2bad96980fbed5b3246e2ba0b195/.github/workflows/ci.yml#L40-L59

🤖 Generated with [Claude Code](https://claude.ai/code)

iammojogo-sudo:
Will look into them today :)

iammojogo-sudo:
Updates in this revision:

Fixed the two issues from the code review:

Gizmo mode re-activation bug — The reset effect now clears gizmoMode when both mesh and point light are deselected (clicking empty space or pressing Escape), preventing the rotate/scale tool from silently re-activating on the next selection.
Out-of-scope changes removed — Rebased onto latest dev with only the point light feature commits. Version bumps, build-macos CI job, and README edits that were already on main have been removed.
Additional improvements made during testing:

Sprite click event fixed — Changed point light sprite from onPointerDown to onClick so stopPropagation() correctly prevents the mesh click handler from firing when selecting a point light.
Keyboard shortcuts work for point lights — R (rotate) and S (scale) keys now work for both mesh and point light selections. Removed the hasModel guard that was blocking shortcuts when no mesh was loaded.
Toolbar shows for point light selection — Move, Rotate, and Scale buttons appear when a point light is selected (not just when a mesh is selected).
Rotate/Scale gizmos on point lights — All three transform gizmos render on point lights for consistency.


Planned future updates (separate PR):

Shadow mapping for point lights — Currently point lights illuminate through meshes because there's no shadow calculation. Enabling castShadow on point lights and configuring shadow maps would make lighting more realistic.
Light source types — Add options for different light types (point, directional, spot) with configurable properties per light.

lightningpixel (approved):

iammojogo-sudo (commented):
I love the updated changes and purple glow idea. :)

PR #233, opened 2026-07-25
fix(extensions): validate model registration

### LINKED ISSUE #232: Model extension install can finish without runtime registration
### Describe the bug
A model extension can be shown as installed while its IDs are absent from the Python registry. Workflows then return `Unknown model ID`, and `/extensions/errors` remains empty.

When an extension with `setup.py` has no usable venv, the registry falls back to direct loading, does not add the extension root to `sys.path`, and drops discovery exceptions. Installer completion also does not validate the expected IDs.

### Steps to reproduce
1. Install a model extension whose root generator imports a sibling package.
2. Interrupt setup or remove its platform venv executable.
3. Start or reload Modly and request its node.

### Expected behavior
The extension requires Repair, discovery errors are exposed, and installation only completes after all expected IDs are registered.

### Actual behavior
Direct discovery fails with `No module named 'pixal3d_extension'`; `load_errors()` is empty; generation returns `Unknown model ID: pixal3d/generate`; a stale `.modly-incomplete` folder can appear as a blank duplicate card.

### Modly version
v0.4.1 and `dev` 088b25b91ec41de8c944bee630140b9c5cf86f99.

### Operating system
Reported on Windows 11; reproduced in an isolated Linux registry harness.

### Screenshots & logs
```text
[Registry] ERROR loading extension 'pixal3d': No module named 'pixal3d_extension'
{"detail":"Unknown model ID: pixal3d/generate"}
```

### Confirmation
- [x] I searched existing issues.
- [x] I reproduced this on the latest dev branch.

DrHepa:
Fixes #232

## Summary

- make model-extension installation success depend on successful runtime registration
- preserve actionable per-extension discovery errors instead of silently dropping failed generators
- make Install, Repair, and local replacement crash-safe with rollback and durable quarantine
- keep incomplete models repairable in the UI while excluding them from generation workflows

## Key changes

- isolates direct Python extension imports and reloads extension-owned modules without leaking extension roots through global `sys.path`
- validates reload results strictly for model extensions while preserving the public reload response and process-extension behavior
- records registration transactions outside extension trees, restores the previous version after failed updates, and prevents uninstall/restart resurrection
- uses exact-destination, one-time validation capabilities for quarantined extensions
- stops loaded extension runtimes before Repair mutates their environment
- fails closed for unreadable or incompatible replacement manifests
- preserves UI-managed Models data; no weights are downloaded, deleted, or moved

## Validation

- [x] `npm test` — 47 Python and 110 Node tests passed in the configured Modly development environment
- [x] `npm run lint`
- [x] `npm run build`
- [x] Python compilation checks
- [x] `git diff --check upstream/dev --`
- [x] Node and web TypeScript diagnostics match the clean `upstream/dev` baseline

### DIFF
--- a/api/routers/extensions.py
+++ b/api/routers/extensions.py
@@ -1,19 +1,24 @@
 import asyncio
 import subprocess
 import sys
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, Body, HTTPException
 
 router = APIRouter(tags=["extensions"])
 
 
 @router.post("/reload")
-async def reload_extensions():
+async def reload_extensions(payload: dict | None = Body(default=None)):
     """
     Re-scans the extensions/ folder and reloads the registry without restarting FastAPI.
     Unloads all currently loaded generators before reloading.
     """
     from services.generator_registry import generator_registry
-    generator_registry.reload()
+    validation_capability = None
+    if isinstance(payload, dict):
+        candidate = payload.get("validationCapability")
+        if isinstance(candidate, dict):
+            validation_capability = candidate
+    generator_registry.reload(validation_capability)
     return {
         "reloaded": True,
         "models":   list(generator_registry._generators.keys()),
--- a/api/services/extension_process.py
+++ b/api/services/extension_process.py
@@ -372,14 +372,24 @@ class ExtensionProcess:
         next load() starts a fresh subprocess.
         """
         proc = self._proc
-        self._proc   = None
         self._loaded = False
         if proc and proc.poll() is None:
             try:
                 proc.kill()
                 proc.wait(timeout=5)
-            except Exception:
-                pass
+            except Exception as exc:
+                # Keep the process reference so callers can verify that the
+                # worker may still be alive and refuse to mutate its venv.
+                self._proc = proc
+                raise RuntimeError(
+                    f"[{self.MODEL_ID}] Could not stop extension subprocess"
+                ) from exc
+            if proc.poll() is None:
+                self._proc = proc
+                raise RuntimeError(
+                    f"[{self.MODEL_ID}] Extension subprocess is still running"
+                )
+        self._proc = None
         self._drain_queue()
 
     def _drain_queue(self) -> None:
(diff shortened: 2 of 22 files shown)


lightningpixel:
Thanks for this, the design looks solid and the test coverage is impressive.

Quick question before I approve: did you actually run a generation with trellis2 on this branch ?

It's the one official extension with no setup.py (just build_vendor.py + a committed vendor/), so after this PR it stays on the direct/proxy path permanently, it can never get a venv.

Two things I couldn't verify by reading:

1. Cancel during generation. call_in_context holds the manager-wide RLock for the whole call. POST /generate/cancel does hasattr(gen, "_proc") on the event loop, which goes through __getattr__ → activate() → waits on that lock. Looks like the API would freeze for the entire generation.
2. Vendored imports. _setup_vendor() inserts vendor/ into sys.path from load(), but activate() restores sys.path on exit. Any lazy import during generate() might then fail.

If you've tested it and both work, let me know and I'll drop it. Otherwise that's the scenario I'd want checked before merge.

DrHepa:
Thanks — I tested the current functional Trellis2 GGUF extension at commit d707382 in a completely isolated environment on PR #233.
This repository does have a setup.py and creates an extension venv. Both trellis2/generate and trellis2/refine therefore register as ExtensionProcess; neither uses the direct proxy/RLock path.

I exercised the real FastAPI reload and cancellation routes. Reload registered both nodes with no errors, and cancellation returned HTTP 200 in 0.86 ms, reached the cancelled state, and produced zero legacy proxy/context activations.

Lazy imports also occur inside the extension subprocess using its own venv, so restoring the manager’s sys.path does not affect this GGUF extension.

I want to be precise about the limit: I could not complete a real GLB generation on this ARM64 GB10 machine. Trellis2’s setup exits 0, but the required cumesh and flex_gemm wheels are unavailable for ARM64. Registration and the worker ready handshake succeed, but generation readiness does not. That appears to be an existing Trellis2 GGUF setup/platform issue rather than a regression from this PR.
Based on the functional GGUF extension path, I could not reproduce either of the two PR concerns.

lightningpixel (approved):

PR #242, opened 2026-07-27
Play animation clips in the 3D viewer

OmarB97:
## What
The viewer creates an `AnimationMixer` for models that carry animation clips, ticks it from `useFrame`, and disposes it with the model. When a file has more than one clip, a small picker lets you switch between them.

## Why
`useGLTF` already parses `gltf.animations`, but nothing ever plays them — a rigged model renders frozen in its bind pose. From the outside that looks like the rig failed, which is a confusing first experience of any rigging extension (UniRig, SkinTokens, or a hand-made rig). Three lines of state and a mixer make the model move.

Tested on Linux with GLB files carrying one clip and three clips (skinned mesh, 8-bone chain): single clips autoplay, multi-clip files default to the first and switch cleanly via the picker, and unloading a model stops and uncaches the mixer.

## Notes
- No new dependencies — `AnimationMixer` comes from the `three` already imported, and `Html` from the `@react-three/drei` already in use.
- Models without clips take an early return, so the static path is unchanged.

### DIFF
--- a/src/areas/generate/components/Viewer3D.tsx
+++ b/src/areas/generate/components/Viewer3D.tsx
@@ -2,7 +2,7 @@ import { Component, Suspense, useCallback, useEffect, useMemo, useRef, useState 
 import type { ReactNode, ErrorInfo, MutableRefObject } from 'react'
 import { Canvas, useFrame, useLoader, useThree } from '@react-three/fiber'
 import type { ThreeEvent } from '@react-three/fiber'
-import { Environment, GizmoHelper, Lightformer, OrbitControls, useGizmoContext, useGLTF } from '@react-three/drei'
+import { Environment, GizmoHelper, Html, Lightformer, OrbitControls, useGizmoContext, useGLTF } from '@react-three/drei'
 import { EffectComposer, Outline, Select, Selection } from '@react-three/postprocessing'
 import * as THREE from 'three'
 import { OBJLoader } from 'three/examples/jsm/loaders/OBJLoader.js'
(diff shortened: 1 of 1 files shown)


lightningpixel:
Thanks for tackling this — animated GLBs rendering frozen in bind pose is a real papercut, and the overall shape of the change (mixer per scene, `useFrame` tick, early return when there are no clips) is the right one.

Before anything else though: **this branch doesn't build.** Line 189 has a JSX expression with no right-hand operand:

```jsx
<option key={clip.name || i} value={i}>{clip.name || }</option>
```

Reproduced on a fresh checkout of the PR head:

```
X [ERROR] Unexpected "}"

    src/areas/generate/components/Viewer3D.tsx:189:67:
      189 │ ...  <option key={clip.name || i} value={i}>{clip.name || }</option>
          ╵                                                           ^
```

Because it's a parse error rather than a runtime one, esbuild bails during transform, so `npm run dev` and `npm run package` both fail and the whole app fails to load — not just the viewer. Looks like a fallback (something like `` `Clip ${i + 1}` ``) got lost in an edit.

Worth double-checking `npm run dev` before the next push — the rest of the review is hard to act on while the branch can't run.

Three other things I noticed while reading, so they can go in the same round:

- **`clipIndex` never resets between models.** `<MeshModel>` is rendered without a `key` and `modelUrl` flips straight from one done-job URL to the next, so React keeps `GltfMeshModel` mounted and the state survives. Go to a 3-clip model, pick clip 3, then switch to a 2-clip model: the mixer plays `animations[Math.min(2, 1)]` — the *second* clip, not the first as the PR describes — and `<select value={2}>` has no matching `<option>`, so the picker renders blank. A `key={modelUrl}` on `MeshModel`, or a reset effect on `scene`, fixes both.

- **`key={clip.name || i}` collides on duplicate names.** Plenty of exporters emit several clips called `Take 001` or `mixamo.com`. `key={i}` is safe here since the list is positional and never reordered.

- **The picker is anchored inside the model.** `<Html position={[0, 0, 0]}>` with no `center` or screen-space anchoring projects the `<select>` at the model's origin, so it sits on top of the mesh and slides off-screen as you orbit. Also, `wrapperClass="clip-picker"` has no matching CSS anywhere in the repo, so that class is currently dead. This probably wants to live in the DOM overlay next to `ViewerToolbar` rather than in the 3D scene.

One thing that is correct and worth noting: the mixer teardown is fine — `stopAllAction()` goes through `_deactivateAction` → `binding.restoreOriginalState()`, so the scene returns to bind pose before the next mixer binds, and switching clips won't leave a stuck pose.

PR #252, opened 2026-07-28
Carry artist intent through generated assets

OmarB97:
## Summary

- create a run-specific `modly.intent.v1` record before the first Style/model stage
- preserve source image, project, chosen Style label and prompt, Extra details, and Movement as distinct exact fields
- carry the record after every successful image or mesh stage, failing loudly if context cannot travel
- merge exporter metadata atomically instead of recreating the final sidecar
- preserve tags and `derived_from` lineage while retaining the full intent history

This is stacked on #251 because it branches from the newest installed desktop build.

## Verification

- `npm run build`
- `npm run lint -- --quiet`
- `npm test` (23 Python tests and 94 Node tests)
- `npm run package -- --linux dir`
- installed the packaged build without restarting the operator's open Modly session
- real desktop reference-image run: Style -> TRELLIS -> rig/animate -> optimizer -> GLB export
- final sidecar retained both exact artist sentences, Style choice, project, geometry findings, operator override evidence, and every carried stage event
- real uncertain run stopped after TRELLIS, persisted `needs_clarification`, and surfaced the exact question in the desktop UI without exporting or guessing
- Library loaded the final rigged asset with tags, thumbnail, and two validated motion previews; catalog grew only by the four generated assets
- compiled exporter probe preserved both the intent record and parent/root lineage


lightningpixel:
Thanks for the volume of work here — and before anything else, credit where it's due: the IPC layer in the Library work is careful. `readWorkspaceAssetLibraryThumbnail` reuses `normalizeWorkspaceAssetPath` rather than rolling its own check, and I verified the extension swap runs on the already-normalized path, so traversal stays closed. That's the right instinct and it's not what I usually see in a first contribution.

Two structural things need sorting before any of this can be reviewed properly.

## 1. These eleven PRs are one stack, and they're all pointed at the same base

#245 through #252 aren't independent changes — they're a single 21-commit chain, each branch built on the previous one, but every PR opened against `main`. So each PR's diff contains all of its predecessors: #243's commit `9112730` is inside #245, #247, #248 … all the way up to #252 here at +5049.

That means merging any one of them merges everything beneath it, and reviewing them separately means reviewing the same code up to eight times. I can't give you useful feedback in that shape.

Please either squash the chain into one PR, or restack it properly — set each PR's `base` to the branch below it, so each diff shows only its own commits. #242 and #244 are genuinely independent and can stay as they are.

I've retargeted all of them from `main` to `dev`, which is the integration branch for this repo. Heads up that #244 and this one now show conflicts against `dev` — it picked up two extension fixes that touch `mockExtensions.ts`, `ipc-handlers.ts`, `index.ts` and `electron-api.ts`.

## 2. Nothing in the repo produces the files this stack consumes

The thumbnail code reads a sibling `<name>.thumb.png`, and the comment points at `thumbs.py`. There is no `thumbs.py` in this repository, and nothing anywhere writes a `.thumb.png` — the only other hit for "thumbnail capture" is `Viewer3D.tsx`, where the body is `captured.current = false` under a "kept for future use" comment.

By the top of the stack this has grown rather than resolved: `INTERNAL_SUFFIXES` now lists `.thumb.png` and `.previews.json`, and there are tests for preview-clip WebP fetching. So there's a full consumption layer for a rendering pipeline that doesn't exist here.

On a real install every one of these lookups returns `{success: false}`, every row renders exactly as it did before, and the only measurable change is an extra IPC round-trip per GLB on each Library refresh. The tests pass because they write the fixture themselves (`writeFile(… 'hero.thumb.png', 'fake-png-bytes')`), so they verify the reader against a file the product never creates.

Where is the producer? If it's in a separate repo or an extension, say so and link it. If it doesn't exist yet, it needs to land first — otherwise this is several thousand lines whose user-visible effect is nil.

## 3. The stack contradicts itself

`28a849c` is titled *"Keep one saved interface scale instead of two that fight"* — it's fixing a conflict created by earlier commits in this same chain (the `modly-library-panel-width` localStorage key from `9112730` against `be4de10`'s remembered options panel). The panel resize introduced in `9112730` is likewise reworked twice more, by `067ca76` and `4d12003`.

That's the clearest sign the chain moved faster than it could be checked. It also means detailed review of the earlier commits is wasted effort — I'd be commenting on code your own later commits already replace.

**What would help most:** pick the one change you're most confident in, restack it alone against `dev`, confirm you've seen it work in the running app, and let's get that merged. Once we have a rhythm on a single reviewable PR, the rest can follow.

PR #267, opened 2026-08-17
Add AMD ROCm support (Linux + Windows)

lionrayonnant:
## Summary

Adds AMD GPU support alongside the existing NVIDIA/MPS/CPU paths. Modly never
installs PyTorch itself — each extension's `setup.py` does, from an index it
hardcodes — so this is really two pieces: detecting AMD hardware, and
redirecting extension installs to the right ROCm wheels.

- **`electron/main/gpu-detect.ts`** (new) — detects AMD GPUs via KFD topology
  on Linux and `Win32_VideoController` on Windows, resolves the ROCm pip index
  and requirements per platform. NVIDIA keeps detection priority, so nothing
  changes for existing CUDA/MPS users. Escape hatches (`MODLY_TORCH_FLAVOR`,
  `MODLY_ROCM_GFX`, `MODLY_ROCM_INDEX`, `MODLY_ROCM_TORCH_SPEC`) are available
  for machines the auto-detection gets wrong, or to roll back to an older
  ROCm/torch pairing.
- **`electron/main/setup-launcher.ts`** (extracted from `ipc-handlers.ts`) —
  adds a compatibility shim that rewrites an extension's pip torch install to
  ROCm wheels. This is needed because most extension `setup.py` scripts
  predate AMD support and hardcode a CUDA index; the official
  `hunyuan3d-mini` extension already accepts a `torch_flavor` argument, but
  Modly never sent it, and other extensions don't know the argument exists
  at all.
- **`api/routers/extensions.py`** — the FastAPI-side GPU detection no longer
  mistakes a ROCm build's device capability for a CUDA one (both answer
  `torch.cuda.get_device_capability()` the same way under PyTorch's HIP
  build), and reads the AMD compute target from KFD topology directly (no
  `torch` dependency in this process).
- **`electron/main/copy-runtime.ts`** (new) — fixes an unrelated but blocking
  bug found while verifying this end-to-end on Linux: `fs.cp` rewrites the
  bundled Python runtime's relative symlinks into absolute paths pointing at
  the *source* tree. On AppImage, the source is the ephemeral
  `/tmp/.mount_Modly-XXXXXX/` mount, so the "stable" runtime copy silently
  wasn't stable — every extension venv recorded that path and died on the
  next launch. `verbatimSymlinks: true` keeps the links relative.
- **`docs/running-on-amd-rocm.md`**, **`arch/decisions/AMD-ROCM-SUPPORT.md`** —
  usage, the verified configuration, escape hatches, and known limitations.

## Verified on real hardware

Radeon RX 9060 XT (Navi 44, gfx1200), Linux:

- Detection reports `accelerator=rocm gfx=gfx1200` correctly.
- `hunyuan3d-mini` extension setup installs `torch 2.13.0+rocm7.2` and loads
  on the GPU (`torch.cuda.is_available()` is `True`, `gcnArchName` is
  `gfx1200`).
- A full image-to-3D generation through the normal `ExtensionProcess`
  subprocess path completes successfully (18s load, 221s generation, valid
  GLB output).
- rocBLAS (fp16 matmul) and MIOpen (conv2d) kernels both execute correctly.
- 14 GB of the card's 16 GB allocates and reads back cleanly —
  [ROCm/ROCm#6295](https://github.com/ROCm/ROCm/issues/6295), which reports
  this exact card capped near 8GB, did not reproduce on this stack.

Windows is implemented (wheel URLs and `cp311` availability were checked
against the AMD and PyTorch package indexes) but not run end-to-end — no
Windows/AMD machine was available for this PR.

## Extensions tried beyond hunyuan3d-mini

Two other official extensions were tried while verifying this, both without success:

- **`modly-trellis2-extension`** — not attempted. Its `build_vendor.py` ships
  precompiled CUDA binaries (`nvdiffrast`, `cumesh`, `spconv`) in `vendor/`,
  built against `nvcc`. Independent of anything in this PR, those `.so` files
  cannot run under ROCm/HIP — there's no ROCm build shipped, and nothing here
  can redirect a vendored binary the way it redirects a pip install.
- **`modly-triposplat-extension`** — attempted and failed. It's pure Python +
  PyTorch (no compiled extensions), so the ROCm redirect in this PR applies
  cleanly and the diffusion stage runs correctly on the GPU. Generation
  consistently fails afterwards, during mesh reconstruction
  (`splat_mesh.py`'s Gaussian-to-mesh step): comparing CPU vs GPU on the same
  captured real output showed the accumulated density values diverge
  significantly (~13x) between devices, though I could not isolate this to
  a single faulty op — `_inverse_covariance` and `index_add_` each checked
  out correctly in isolation with the same data. This looks like a
  ROCm-specific numerical issue somewhere in that extension's reconstruction
  pipeline (torch 2.13.0+rocm7.2), not something this PR's detection/redirect
  layer can fix. Flagging it here rather than silently leaving it unverified.

Extensions with vendored CUDA-compiled native code (e.g. `nvdiffrast`,
`spconv` binaries built against `nvcc`) cannot work under ROCm regardless of
this change — there's no ROCm equivalent shipped. Pure-PyTorch extensions
are the ones this PR actually helps; `hunyuan3d-mini` is confirmed working
end-to-end, `triposplat` is not (see above).

## Test plan

- [x] `npm run lint`
- [x] `./node_modules/.bin/tsc --noEmit -p tsconfig.node.json` (no new errors)
- [x] `npm test` (23 Python + 77 Node tests, including 25 new tests for the
      detection/redirect logic added here)
- [x] End-to-end on real AMD hardware (see above)


lightningpixel (changes requested):
Thanks a lot for this — ROCm support is something we've wanted for a while, and reading the KFD topology instead of asking torch is the right call.

I've reviewed the diff in detail. Two things to do before this can land:

## 1. Resolve the conflicts

I've retargeted the PR to `dev` — that's where feature work lands, `main` only receives releases. It now conflicts on `api/routers/extensions.py` and `electron/main/ipc-handlers.ts`, so please merge `dev` into your branch first. Those are also the two files carrying the most important issues below, so it's worth doing both in one pass.

## 2. Then fix the issues below

### Blocking

**a. `npm test` fails on Windows** — `electron/main/copy-runtime.test.mjs:38` calls `symlinkSync` with no platform guard, and `npm run test:node` picks the file up unconditionally. On Windows symlinks need Developer Mode or an elevated shell, so it throws `EPERM` (77 tests, 3 fail on a normal Windows dev machine). Windows is the primary dev platform here. `setup-launcher.test.mjs` already handles this with `{ skip: !PYTHON }` — please do the same, e.g. `{ skip: process.platform === 'win32' }`.

**b. `_detect_gpu_sm()` always returns 0** — `api/routers/extensions.py:95` does `import torch`, but this process runs in Modly's main venv, which has no torch (`api/requirements.txt`). Your own docstring on `_detect_gfx_target` says exactly this. So at lines 60-61 both `accelerator` and `torch_flavor` resolve to `"cpu"` on every NVIDIA machine going through this endpoint: an RTX 4090 user installing via `POST /extensions/setup/{id}` gets CPU-only torch and then generates on CPU, silently. Needs a torch-free detection path (parsing `nvidia-smi`, like `gpu-detect.ts` does), or the endpoint should decline to guess rather than claim `cpu`.

**c. The `setup.py` JSON contract differs between its two producers** — `extensions.py:53-63` and `ipc-handlers.ts:64-78` both build the arguments handed to `setup.py`, but the Python side omits `arch` and `torch_index_url`. A `setup.py` reading `args["arch"]` — exactly what the aarch64 wheel path covered in `setup-launcher.test.mjs` does — raises `KeyError` and the endpoint turns it into a 500. Both producers need to emit the same keys; a test asserting that would be welcome.

**d. ROCm/CUDA priority is inverted between the two paths** — `extensions.py:60` gives ROCm priority over CUDA, while `gpu-detect.ts` deliberately keeps NVIDIA first. Any Ryzen with an integrated Radeon publishes a KFD node with `simd_count > 0`, so a Ryzen + NVIDIA dGPU box resolves `rocm` here and `cuda` on the Electron path — same machine, two different torch stacks depending on which code ran setup. Please match `gpu-detect.ts`.

### Should fix

**e. `_is_pip_command` misses a common pip invocation** — `setup-launcher.ts:78` only inspects `command[:3]`, so `[sys.executable, "-u", "-m", "pip", "install", "torch"]` isn't recognised and the whole ROCm redirect is skipped. Silently: no log line, exit 0, CUDA torch on an AMD machine. Scan the full command for the `pip` / `-m pip` token.

**f. A foreign `--index-url` survives the rewrite and wins** — in `_rewrite_rocm` (`setup-launcher.ts:162`), an `--index-url` that is neither a `download.pytorch.org/whl/` URL nor matched by `_is_rocm_index` is kept *and* the ROCm one is injected too. pip is last-wins, so `pip install torch trimesh --index-url https://pypi.org/simple` ends up installing from PyPI while the log claims the redirect succeeded. Note `_is_rocm_index` looks for `/whl/rocm` and therefore doesn't match our own Windows index `https://repo.amd.com/rocm/whl-multi-arch/`.

**g. The subprocess shim has gaps the ROCm path now depends on** — `setup-launcher.ts:225` patches only `run`, `check_call` and `check_output`. `subprocess.Popen([...])`, or passing the command as the `args=` keyword (the `if args:` guard in `_patched_run` skips it), bypasses the redirect entirely. This was low-stakes when the shim only stripped a macOS index; now every AMD install rides on it.

**h. The PyPI rescue index is missing on one branch** — `setup-launcher.ts:194` adds `--extra-index-url https://pypi.org/simple` only in the `not keeps_rocm_index` branch. An extension already on a ROCm index (hunyuan3d-mini's Linux ROCm branch) installing torch alongside app packages sends them out with only the ROCm index in play, and `trimesh`/`diffusers` 403 — the exact failure the comment on line 191 says the code exists to prevent.

**i. First-KFD-node-wins picks the wrong GPU on APU + dGPU systems** — `parseKfdGfxTarget` (`gpu-detect.ts:130`) takes the first node with `simd_count > 0`, sorted by node number, and the APU is commonly the lower one. A Ryzen 7840 + RX 7900 XTX resolves `gfx1103` instead of `gfx1100`: on Windows that pins a `torch[device-gfx1103]` AMD doesn't publish, on Linux the ROCm wheels carry no gfx1103 kernels, so `torch.cuda.is_available()` is `False` after a multi-GB download. Prefer the discrete node (or the highest `simd_count`). Related: APU-only Linux machines that used to get CPU torch now get ROCm wheels they can't use.

### Minor

**j. `MODLY_TORCH_FLAVOR` doesn't override as documented** — the log says "skipping auto-detection" and `docs/running-on-amd-rocm.md:40` says an explicit override wins over everything, but the darwin/arm64 check at `gpu-detect.ts:355` runs first and returns `mps`, and elsewhere `nvidia-smi` is still probed; when it fails, forcing `cuda` yields `{sm: 0, cudaVersion: 0}` and a quiet CPU install. Either short-circuit properly or soften the docs.

**k. `queryWindowsVideoControllers()` runs twice** — on Windows with no mapped AMD card it's spawned at `gpu-detect.ts:384` and again at line 372 purely to build the log message: two PowerShell + CIM round-trips (~1s each) on every install and repair. Reuse the first result.

---

For the record, these all checked out fine: `formatGfxTarget` decodes correctly on gfx90a / gfx942 / gfx908 / gfx1030 / gfx1200 / gfx1201 and the `minor > 15 || step > 15` guard is right; `_TORCH_REQ_RE` doesn't false-match `torchsde` or `torch-scatter`; the `verbatimSymlinks: true` fix in `copyRuntimeTree` is a good catch; the `torch.version.hip` guard is safe on CPU-only builds.

Ping me once the conflicts are resolved and the items above are addressed and I'll take another pass. Happy to help with any of them — particularly b, c and d.

lightningpixel (approved):

PR #275, opened 2026-08-21
feat(models): support multiple Hugging Face sources per node

### LINKED ISSUE #273: Support multiple Hugging Face model sources per node
## Problem

A model node can currently declare only one `hf_repo` and one `download_check`. Models that depend on weights from multiple Hugging Face repositories must implement custom bootstrap or runtime downloads, so those assets cannot be installed, monitored, paused, resumed, repaired, or removed consistently from the Modly Models/Extensions UI.

## Proposed solution

Add a backwards-compatible `model_sources` array to model nodes. The first version would support Hugging Face sources only.

Each source should declare:

- A stable source ID.
- A Hugging Face repository ID and optional revision.
- A destination relative to the node model directory.
- Optional include/skip filters.
- One or more files used to verify that the source is complete.

Example:

```json
{
  "model_sources": [
    {
      "id": "primary",
      "provider": "huggingface",
      "repo_id": "org/main-model",
      "destination": ".",
      "checks": ["model_index.json"]
    },
    {
      "id": "encoder",
      "provider": "huggingface",
      "repo_id": "org/encoder",
      "destination": "auxiliary/encoder",
      "include_prefixes": ["config.json", "model.safetensors"],
      "checks": ["config.json", "model.safetensors"]
    }
  ]
}
```

The host should:

- Normalize legacy `hf_repo` / `download_check` fields into a single source, preserving existing manifests.
- Resolve the download plan from the installed manifest instead of accepting arbitrary repositories or destinations from the renderer.
- Confine every destination and check to the node model directory.
- Download all sources through one node action with aggregate progress and existing pause/resume/cancel controls.
- Mark the node installed only when all source checks pass.
- Remove all node-owned sources when model weights are deleted.

## Alternatives considered

Making `hf_repo` an array leaves destinations, per-repository filters, and readiness checks ambiguous. Keeping downloads inside extension setup/runtime code works for individual extensions, but bypasses the common model lifecycle and creates inconsistent offline behavior.

## Additional context

This proposal deliberately limits the first version to Hugging Face repositories. Other providers or arbitrary URLs can be considered separately with their own integrity and security requirements.

## Checklist

- [x] I searched for an existing request before opening this issue.

DrHepa:
Summary

- add an optional, backwards-compatible "model_sources" contract for model nodes that require weights from multiple Hugging Face repositories
- resolve download plans from the installed manifest in Electron main and download all declared sources sequentially through one aggregate action
- validate source destinations, filtered plans, and destination collisions before writing model data
- require each readiness check to be a regular, non-empty file; zero-byte final files are replaced instead of being treated as installed
- aggregate progress and apply the existing pause/resume/cancel controls to the whole node action
- keep all downloaded sources owned by the existing node model directory, including partial-data cleanup and uninstall handling

Legacy "hf_repo", "download_check", and filter fields keep their existing behavior when "model_sources" is absent.

Provider scope

This PR remains intentionally limited to Hugging Face. GitHub repository archives and GitHub Release assets have different versioning, authentication, rate-limit, redirect, and integrity semantics, so a future "github_release" provider should be designed and reviewed separately instead of adding generic GitHub URLs to this contract.

Testing

- "npm test" — 77 Python tests and 91 Node tests passed
- "./node_modules/.bin/tsc --noEmit"
- "npm run build" — builtin extensions and Electron renderer/main/preload built successfully
- "python -m compileall -q api"
- "git diff --check"
- live two-source Hugging Face smoke test using "hf-internal-testing/tiny-random-bert" and "hf-internal-testing/tiny-random-gpt2"; both files were downloaded to their declared destinations, progress stayed monotonic, and composite readiness passed

Remaining platform validation

- a packaged Windows UI E2E was not run in this Linux environment

Closes #273

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,42 @@ Modly supports external model and process extensions. Each extension is a GitHub
 
 ![Install models](docs/install-models.png)
 
+### Multiple Hugging Face repositories per model node
+
+A model node whose weights are split across repositories can declare
+`model_sources`. Modly validates every source, downloads them sequentially in
+one Models-page action, and considers the node installed only when every
+declared check exists.
+
+```json
+{
+  "id": "generate",
+  "model_sources": [
+    {
+      "id": "primary",
+      "provider": "huggingface",
+      "repo_id": "org/main-model",
+      "destination": ".",
+      "checks": ["model.safetensors"]
+    },
+    {
+      "id": "encoder",
+      "provider": "huggingface",
+      "repo_id": "org/encoder",
+      "revision": "v1.0",
+      "destination": "auxiliary/encoder",
+      "include_prefixes": ["config.json", "model.safetensors"],
+      "checks": ["config.json", "model.safetensors"]
+    }
+  ]
+}
+```
+
+`destination`, filters, and checks use safe POSIX paths relative to the node's
+model directory. The only supported provider is `huggingface`. Existing nodes
+that use `hf_repo`, `download_check`, `hf_include_prefixes`, and
+`hf_skip_prefixes` keep their original behavior.
+
 ---
 
 ## Workflows
--- a/api/routers/model.py
+++ b/api/routers/model.py
@@ -8,9 +8,16 @@ from pathlib import Path
 from typing import Optional
 from urllib.error import HTTPError, URLError
 from urllib.request import Request, urlopen
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, HTTPException, Request as FastAPIRequest
 from fastapi.responses import StreamingResponse
 from services.generator_registry import generator_registry, MODELS_DIR
+from services.model_sources import (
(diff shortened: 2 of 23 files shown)


lightningpixel:
Pushed two small fixes on top of this:

- `hf-download-sources` never actually returned the "sources are required" error when the field was omitted — `normalize_model_sources` always saw the key present, so its own less-clear message fired first.
- `model:cancelDownload` was wiping the whole model directory on cancel. That was fine for a single all-or-nothing repo, but now that a node can have multiple `model_sources` sharing one directory, cancelling while one source is still downloading was deleting already-completed sources too. It now only removes in-progress `.part` files, matching the backend's own cancel cleanup.

Ran the full test suite (Python + Node) and `tsc --noEmit` after — all green.

lightningpixel (approved):

PR #286, opened 2026-08-30
fix(agent): surface HTTP errors from the unload_models tool

kevin9327:
## What

The `unload_models` chat tool discarded the response from `POST /model/unload-all` and unconditionally returned the success string, so a *failed* unload (any 4xx/5xx) still told the assistant — and the user — that VRAM had been freed.

## Why it triggers

`execute_tool` (`api/routers/agent.py`) wraps every tool call in a shared `except httpx.HTTPStatusError` handler that turns an HTTP error status into `"API error <status>: ..."`. But httpx does **not** raise on an error status unless you call `raise_for_status()`. Every other POST-based tool — `list_models`, `decimate_mesh`, `smooth_mesh`, `get_generation_status` — captures the response and calls `raise_for_status()`, and the MCP server's `modly_unload_models` does too. `unload_models` was the only one that fired the request and threw the response away, so a 500 from the unload endpoint was reported to the model as a success.

```python
elif name == "unload_models":
    await client.post(f"{MODLY_API}/model/unload-all")   # response ignored
    return "All 3D generation models have been unloaded from VRAM.", None
```

## Fix

Capture the response and call `raise_for_status()`, matching the sibling tools and the MCP server. The existing `HTTPStatusError` handler then surfaces the real failure.

## Verification

- Added `api/tests/test_agent_router.py` (unittest + `httpx.MockTransport`). `test_http_error_is_surfaced` **fails before** this change (it gets the success string back for a mocked 500) and **passes after**; `test_success_still_reports_unloaded` guards the happy path.

### DIFF
--- a/api/routers/agent.py
+++ b/api/routers/agent.py
@@ -267,7 +267,8 @@ async def execute_tool(name: str, arguments: dict, context: dict) -> tuple[str, 
                 return f"Available models:\n{lines}", None
 
             elif name == "unload_models":
-                await client.post(f"{MODLY_API}/model/unload-all")
+                r = await client.post(f"{MODLY_API}/model/unload-all")
+                r.raise_for_status()
                 return "All 3D generation models have been unloaded from VRAM.", None
 
             elif name == "get_mesh_info":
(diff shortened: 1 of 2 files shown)


lightningpixel:
### Code review

lightningpixel (approved):

PR #287, opened 2026-08-30
fix(generate): read WORKSPACE_DIR dynamically so relocated output isn't lost

kevin9327:
## What

After the workspace is relocated at runtime (`POST /settings/paths`), 3D generations kept writing their output into the **previous** workspace directory, while `/workspace/{path}` served files from the **new** one — so the finished mesh 404s and never appears in the app.

## Why it triggers

`api/routers/generation.py` bound the workspace path with `from services.generator_registry import WORKSPACE_DIR`, capturing the `Path` at import time. `POST /settings/paths` → `generator_registry.update_paths()` **rebinds** the `WORKSPACE_DIR` module global (and repoints every loaded generator's `outputs_dir`):

```python
# services/generator_registry.py
_self_module.WORKSPACE_DIR = workspace_dir
for gen in self._generators.values():
    gen.outputs_dir = workspace_dir
```

Because `_run_generation` used the stale imported name, it recomputed `coll_dir = WORKSPACE_DIR / collection` against the **old** directory and set `gen.outputs_dir` back to it — overriding the value `update_paths` had just written — then built `output_url` relative to the same stale path. `/workspace/...` reads the *current* `WORKSPACE_DIR` dynamically, so it can't find the file.

This affects both `/generate/from-image` and `/workflow-runs/from-image` (both funnel through `_run_generation`).

## Fix

Reference the module attribute `registry.WORKSPACE_DIR` at call time instead of a name captured at import — the same way `optimize.ply_to_splat` (comment: *"workspace dir may change at runtime"*) and the settings router already read it.

## Verification

- Added `api/tests/test_generation_router.py`: drives `_run_generation` with a fake generator after relocating the workspace, and asserts the output is filed under the **current** workspace (and that `output_url` matches). Fails before, passes after.

### DIFF
--- a/api/routers/generation.py
+++ b/api/routers/generation.py
@@ -9,7 +9,11 @@ from fastapi import APIRouter, File, Form, UploadFile, HTTPException, Background
 from services.generators.base import smooth_progress, GenerationCancelled
 
 import re as _re
-from services.generator_registry import generator_registry, WORKSPACE_DIR
+# Import the module (not the name) so WORKSPACE_DIR is read at call time: the
+# settings endpoint rebinds it when the user relocates the workspace, and a
+# binding captured at import would keep writing output to the old directory.
+import services.generator_registry as registry
+from services.generator_registry import generator_registry
 from schemas.generation import JobStatus
 
 router = APIRouter(tags=["generation"])
@@ -70,7 +74,9 @@ def sanitize_collection(collection: str) -> str:
         return "Default"
 
     try:
-        (WORKSPACE_DIR / collection).resolve().relative_to(WORKSPACE_DIR.resolve())
+        (registry.WORKSPACE_DIR / collection).resolve().relative_to(
+            registry.WORKSPACE_DIR.resolve()
+        )
     except (OSError, ValueError):
         return "Default"
 
@@ -206,7 +212,7 @@ async def _run_generation(job_id: str, image_bytes: bytes, params: dict, collect
             return
 
         # Direct output to the collection subfolder
-        coll_dir = WORKSPACE_DIR / collection
+        coll_dir = registry.WORKSPACE_DIR / collection
         coll_dir.mkdir(parents=True, exist_ok=True)
         gen.outputs_dir = coll_dir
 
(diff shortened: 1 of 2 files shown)


lightningpixel:
### Code review

Found 1 issue:

1. Merging as-is will break `/generate/from-image` with a `NameError` on every request. The PR's branch was cut before `sanitize_collection()` was added to `dev`; that function still references the bare name `WORKSPACE_DIR`, which this PR removes from the module's import (`from services.generator_registry import generator_registry, WORKSPACE_DIR` → `import services.generator_registry as registry` / `from services.generator_registry import generator_registry`). `sanitize_collection` is called unconditionally at the top of `generate_from_image`, so the endpoint would raise `NameError: name 'WORKSPACE_DIR' is not defined` on line 73 for every call. The PR's own new test only exercises `_run_generation` directly and doesn't go through `sanitize_collection`, so it doesn't catch this.

https://github.com/lightningpixel/modly/blob/d4d5fc3a7ea05f4c507144d7496520526cbf7e4b/api/routers/generation.py#L72-L74 (current dev, showing the bare `WORKSPACE_DIR` usage inside `sanitize_collection` that this PR's import change would orphan)

https://github.com/lightningpixel/modly/blob/d4d5fc3a7ea05f4c507144d7496520526cbf7e4b/api/routers/generation.py#L96-L98 (current dev, showing `sanitize_collection` is called unconditionally on every request)

Fix: rebase onto current `dev` and update `sanitize_collection`'s two `WORKSPACE_DIR` references (and its docstring) to `registry.WORKSPACE_DIR`, consistent with this PR's own stated rule in the import comment.

kevin9327:
Rebased onto current `dev` and addressed the review: `sanitize_collection()` now reads `registry.WORKSPACE_DIR` as well (its containment check runs at the top of `generate_from_image`), so no bare `WORKSPACE_DIR` remains in the module. Added `GenerateFromImageWorkspaceTests`, which drives `generate_from_image` itself after relocating the workspace -- on the rebased branch without the fix it fails with `NameError: name 'WORKSPACE_DIR' is not defined`, and passes with it.

Verification (venv with fastapi/python-multipart/httpx, run from `api/`): `python -c "import routers.generation, routers.workflow_runs"` OK; `python -m unittest discover -s tests` -> Ran 69 tests, OK (skipped=2).

lightningpixel (approved):

PR #288, opened 2026-08-30
fix(workflow-runs): take part in the job TTL purge so the headless API doesn't leak

kevin9327:
## What

The headless `/workflow-runs` API leaks job records. It shares the `_jobs`, `_cancel_events` and `_completed_at` dicts with the `/generate` endpoints, but never took part in their TTL purge:

- **`create_run_from_image` never called `_purge_old_jobs()`.** `/generate/from-image` purges terminal jobs older than `_JOB_TTL` on every call; `/workflow-runs/from-image` didn't, so records only ever got swept if a `/generate` call happened to run the purge. For a pure automation client (the surface `/workflow-runs` exists for), nothing triggers it and `_jobs` grows without bound.
- **`cancel_run` never stamped `_completed_at`.** `_purge_old_jobs()` only sweeps entries that have a completion timestamp, and `cancel_job` (the `/generate` sibling) sets one when it cancels. `cancel_run` set `status="cancelled"` but no timestamp, so a cancelled run could **never** be purged — a permanent leak of one `JobStatus` + one `threading.Event` per cancellation.

## Fix

Mirror the `/generate` endpoints: call `_purge_old_jobs()` when a run is created, and record `_completed_at[run_id]` when a run is cancelled. Collection routing is intentionally left untouched.

```python
# create_run_from_image
_purge_old_jobs()
_jobs[job_id] = JobStatus(...)

# cancel_run
if job.status in ("pending", "running"):
    job.status = "cancelled"
    _completed_at[run_id] = time.monotonic()
```

## Verification

- Added `api/tests/test_workflow_runs_lifecycle.py`: one test proves a stale terminal job is evicted when a new run is created, the other proves a cancelled run gets a completion stamp so the purge can reach it. Both **fail before**, **pass after**.

### DIFF
--- a/api/routers/workflow_runs.py
+++ b/api/routers/workflow_runs.py
@@ -1,5 +1,6 @@
 import json
 import threading
+import time
 import uuid
 from typing import Optional
 from fastapi import APIRouter, BackgroundTasks, File, Form, HTTPException, UploadFile
@@ -9,7 +10,9 @@ from routers.generation import (
     VALID_REMESH_MODES,
     _cancel_events,
     _cancelled,
+    _completed_at,
     _jobs,
+    _purge_old_jobs,
     _run_generation,
     sanitize_collection,
 )
@@ -75,6 +78,8 @@ async def create_run_from_image(
     job_id = str(uuid.uuid4())
     image_bytes = await image.read()
 
+    _purge_old_jobs()
+
     _jobs[job_id] = JobStatus(job_id=job_id, status="pending", progress=0)
     _cancel_events[job_id] = threading.Event()
 
@@ -115,6 +120,7 @@ async def cancel_run(run_id: str):
         _cancel_events[run_id].set()
     if job.status in ("pending", "running"):
         job.status = "cancelled"
+        _completed_at[run_id] = time.monotonic()
 
     try:
         gen = generator_registry._generators.get(generator_registry._active_id)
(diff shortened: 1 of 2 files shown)


lightningpixel:
### Code review

lightningpixel:
This branch has conflicts with `dev` that need to be resolved before merging. In particular, the import block in `api/routers/workflow_runs.py` (`from routers.generation import (...)`) diverges from `dev`, which added `VALID_REMESH_MODES` and `sanitize_collection` to that same import. When resolving, make sure both are kept alongside this PR's `_completed_at` and `_purge_old_jobs` additions — dropping either would cause a `NameError` at runtime.

kevin9327:
Rebased onto current `dev` and resolved the import block in `api/routers/workflow_runs.py`: it now carries dev's `VALID_REMESH_MODES` and `sanitize_collection` alongside this PR's `_completed_at` and `_purge_old_jobs` (all eight imported names are used in the module). One test adjustment: the lifecycle test now passes `collection="Default"` explicitly, as the existing router tests do, because dev's new `collection: str = Form("Default")` parameter is a `Form` object when the handler is called directly.

Verification (venv with fastapi/python-multipart/httpx, run from `api/`): `python -c "import routers.generation, routers.workflow_runs"` OK; `python -m unittest discover -s tests` -> Ran 69 tests, OK (skipped=2). Diff vs `dev` is +6 lines in the router plus the new test file.

lightningpixel (approved):

PR #292, opened 2026-09-02
fix: make all typecheck and test gates green on macOS

VanniMnemonic:
## Summary
- Repair the TypeScript gates: enable `rewriteRelativeImportExtensions` (node:test needs `.ts` specifiers for native type-stripping), include `src/shared/types` in the node project, drop the orphaned `collections.ts`, type `Dirent[]` entries, annotate `assetLibraryUi` map callbacks, add a null guard in the UI test.
- Fix a macOS-only bug in `generator_registry`: `os.path.abspath` does not resolve `/var` → `/private/var`, so correctly-authorized extensions stayed blocked as "incomplete installation". Now compares against `ext_dir.resolve()`.
- Make `run-pytests.mjs` prefer `api/.venv` (system Python lacks fastapi) and bump Electron to `^44.1.1` (pending from the packaging work).

## Test plan
- [x] `npx tsc --noEmit` green on all four projects (node, web, root, builtins) — 26 errors → 0
- [x] `npm run test:node` — 118/118 pass
- [x] `npm run test:py` — 63/63 pass (was 2 import errors + 2 failures)
- [x] `npx eslint .` — clean
- [x] `npm run build` — exit 0


lightningpixel:
The macOS symlink fix in `_discover_extensions` looks like it'll break locally-linked extensions:

`registration_authorization[2] == ext_dir.resolve()` now follows the symlink/junction created by `installFromLocal` all the way to its target (the dev's real folder), while `registration_authorization[2]` itself is built with `os.path.abspath()` (which doesn't follow it). They'll never match, so any locally-linked extension gets flagged as "incomplete installation or interrupted runtime registration" right after a successful install — on Windows too, since junctions are resolved by `Path.resolve()` the same way.

Suggest resolving only the parent instead of the full path, mirroring how `destination_path` is built:
```python
== ext_dir.parent.resolve() / ext_dir.name
```
That keeps the `/var` → `/private/var` fix without breaking symlinked/junctioned extension folders. Worth adding a test with a symlinked `ext_dir` since the current suite doesn't cover it.

https://github.com/lightningpixel/modly/blob/f3c1b42c3249c1b4c078796a3a4239ed204f5e98/api/services/generator_registry.py#L457-L465

PR #334, opened 2026-09-08
vram indicator on linux and windows

securitybrahh:
through nvidia-smi


iammojogo-sudo:
I have requested this before also since we already have RAM indicators. I keep opening up task manager to keep an eye on it. Maybe this would be a good feature that @lightningpixel would let me work on?

PR #347, opened 2026-09-14
feat: add "Open in OrcaSlicer" export action

### LINKED ISSUE #346: Add "Open in OrcaSlicer" — one-click slicer hand-off for generated models
Modly generates meshes for 3D printing, but getting one into a slicer today means manually exporting an STL and importing it by hand.

**Proposal:** an **Open in OrcaSlicer** item in the Export dropdown that converts the current model to STL and hands it to OrcaSlicer via its `orcaslicer://open?file=<url>` deeplink, served from the local backend.

**Why a deeplink + STL:** OrcaSlicer registers `orcaslicer://open?file=<url>`, downloads the URL, and derives the import format from the URL's final path segment — and it cannot import GLB. So the model is converted to STL on the fly and served at a path-only URL ending in `model.stl`.

**Scope (v1):**
- STL only (universal; OrcaSlicer auto-repairs). 3MF is a possible fast-follow.
- Reorient Y-up → Z-up and normalize to a sane print size so it imports upright and visible.
- Graceful fallback when OrcaSlicer is not installed.

Happy to implement.

wenghaishi:
One-click hand-off from a generated model to OrcaSlicer via its `orcaslicer://open?file=<url>` deeplink, available in the **Export** dropdown.

let me know if this is alright or there might be some better place for this button.

Screenshot 2026-09-14 at 4 48 50 PM

## What
- **Backend** — new `GET /export/slicer/{fmt}/{token}/model.{fmt}` route that converts the workspace GLB to STL on the fly: bakes scene-graph transforms, reorients **Y-up → Z-up**, and normalizes the longest edge to a sane print size. The URL is intentionally path-only and ends in `model.stl` (no query string), because OrcaSlicer derives the import filename/format from the URL's final path segment. Uses ancestry-based path containment (not string-prefix).
- **Electron** — a `slicer:open` IPC that fires the deeplink and reports success/failure so the UI can fall back when OrcaSlicer isn't installed.
- **Frontend** — an **Open in OrcaSlicer** item in the Export dropdown, shown only for sliceable workspace meshes, plus a pure, unit-tested deeplink builder.

## Why STL + deeplink
OrcaSlicer registers `orcaslicer://open?file=<url>`, downloads the URL, and imports it — and it cannot import GLB. So we convert to STL (which OrcaSlicer auto-repairs) and serve it at a deeplink-shaped URL.

## Testing
- `api/tests/test_export_router.py` — conversion, Y→Z reorientation, print-size normalization, and guards (unsupported format, extension mismatch, malformed token, path traversal, sibling-prefix escape, 404).
- `src/areas/generate/orcaSlicerLink.test.ts` — exact deeplink string, no query string, url-safe base64 round-trip.
- `npm run lint` clean; `npm run test:node` green. Manual E2E verified on macOS (imports upright, ~50 mm).

## Notes / follow-ups
- STL-first by design; 3MF (units + colour) is a natural fast-follow.
- "Not installed" handling is a best-effort error toast; OS-level detection could be added later.

Closes #346


lightningpixel:
Thanks for this — the path containment in the slicer route is solid, and the tests covering the `..` and sibling-prefix escapes are exactly right.

I pushed a commit on your branch. It changes three things:

- **Imported meshes are now sliceable.** `Import` serves files through `/optimize/serve-file`, so `canOpenInOrcaSlicer` hid the action for them — which happens to be the most direct "I have a model, slice it" path. Rather than widening the route's path guard, the backend now keeps an exact-membership registry of the files the user picked themselves this session.
- **Rotation only for glTF sources.** STL/OBJ/PLY are already Z-up, and import converts them to GLB without touching the axes, so the original extension decides. Otherwise an upright STL lands in OrcaSlicer on its side.
- **Rescaling only for unit-sized meshes.** A mesh that already carries a real-world size is the user's own — silently shrinking a 180 mm part to 50 mm would waste a print.

I also dropped the "make sure it is installed" error: on Windows an unregistered scheme still makes `ShellExecuteEx` succeed, so that branch could never run.

Worth flagging: I changed `test_normalizes_longest_edge_to_default_print_size`, since its fixture encoded the unconditional rescale. It's now a unit-sized mesh, matching what the generators actually emit.

lightningpixel (approved):

PR #348, opened 2026-09-14
feat(models): support extension-scoped shared weight groups

### LINKED ISSUE #343: feat(models): support extension-scoped shared weight groups
Summary

Add an opt-in manifest contract that allows multiple model nodes inside the same extension to reuse one or more shared weight groups while retaining their own node-specific weights.

The first phase should be strictly extension-scoped. Sharing weights between unrelated extensions can be designed later once the local ownership, runtime, download, readiness, and deletion semantics are stable.

Problem

Modly currently treats each model capability as both:

1. a capability identity: "<extension-id>/<node-id>"
2. a physical weight owner: "<models-dir>/<extension-id>/<node-id>"

This works for single-node extensions, but a multi-node extension downloads another copy of the same checkpoint for every node, even when those nodes use the same underlying model.

The recently added "model_sources" contract supports multiple Hugging Face repositories per node, but every source is still confined to that node's model directory.

This causes substantial duplication in real extensions:

- SenseNova Vision 7B MoT
  - The extension exposes 15 model nodes.
  - Every node uses the same "sensenova/SenseNova-Vision-7B-MoT" checkpoint and repeats the same download plan.
  - The checkpoint is approximately 29.6 GB.
  - Installing all 15 capabilities can therefore occupy approximately 444 GB instead of one 29.6 GB copy.
- DreamStyle3D
  - "trellis-image-large" and "dreamstyle3d" both need common TRELLIS and DINOv2 files.
  - Approximately 2.85 GB of checkpoint data is currently duplicated.
  - The DreamStyle node also needs private DreamStyle and CLIP weights, so redirecting the entire node directory to another node is not sufficient.
- Pixal3D and a possible WorldSculpt node
  - WorldSculpt uses Pixal3D as its pretrained base and adds its own multi-view conditioning and LoRA/adaptation weights.
  - The official installation downloads both "TencentARC/Pixal3D" and "AlayaLab/WorldSculpt".
  - A WorldSculpt capability inside the Pixal3D extension should reuse the existing Pixal3D weights and download only the additional WorldSculpt assets.

Proposed solution

Introduce extension-scoped shared weight groups.

A shared group owns a validated "model_sources" plan once. Nodes reference the groups they require and may continue declaring private "model_sources".

Possible manifest shape:

{
  "id": "pixal3d",
  "type": "model",

  "weight_groups": [
    {
      "id": "pixal3d-base",
      "model_sources": [
        {
          "id": "pixal3d",
          "provider": "huggingface",
          "repo_id": "TencentARC/Pixal3D",
          "revision": "<pinned-revision>",
          "destination": ".",
          "checks": [
            "pipeline.json"
          ]
        }
      ]
    }
  ],

  "nodes": [
    {
      "id": "generate",
      "weight_groups": [
        "pixal3d-base"
      ]
    },
    {
      "id": "worldsculpt",
      "weight_groups": [
        "pixal3d-base"
      ],
      "model_sources": [
        {
          "id": "worldsculpt-adapter",
          "provider": "huggingface",
          "repo_id": "AlayaLab/WorldSculpt",
          "revision": "<pinned-revision>",
          "destination": "worldsculpt",
          "checks": [
            "worldsculpt/<required-file>"
          ]
        }
      ]
    }
  ]
}

The exact field names are open to adjustment, but shared and node-private sources need separate ownership and storage roots.

Storage contract

Suggested layout:

<models-dir>/
  <extension-id>/
    _shared/
      <weight-group-id>/
        ...
    <node-id>/
      ...

Rules:

- Shared group sources are stored under "_shared/<weight-group-id>".
- Node-specific sources remain under "<node-id>".
- Group IDs are local to their extension.
- Group IDs, destinations, and checks use the same portable-ID and path-confinement rules as "model_sources".
- Absolute paths, "..", symlink escapes, separators inside IDs, and cross-extension references are rejected.
- No symlinks or hard links are required.
- "_shared" should be reserved and rejected as a node ID.

Download behavior

Installing a node should build one effective plan from:

1. all referenced shared weight groups;
2. the node's private "model_sources".

Expected behavior:

- A completed shared group is not downloaded again for another node.
- "Install all" deduplicates shared groups before starting downloads.
- Concurrent node installations cannot write to the same shared group simultaneously.
- Active download locking is based on the canonical physical target, not only the requesting node ID.
- Pause, resume, cancel, repair, progress, revision pinning, filters, and checks continue to use the existing "model_sources" behavior.
- Failure of a node-private source does not mark the shared group as incomplete.
- Partially downloaded shared data remains recoverable through the existing resume/repair flow.

Readiness behavior

A node is ready only when:

- every referenced shared group passes all its checks; and
- every node-private source passes all its checks.

This allows the following states:

- Pixal3D base downloaded → the regular Pixal3D node is ready.
- Pixal3D base downloaded but WorldSculpt weights missing → the WorldSculpt node is not ready.
- Installing WorldSculpt downloads only its missing private sources.
- One SenseNova shared group downloaded → all SenseNova capability nodes that have no private sources become ready.

The Models UI should make this relationship visible, for example:

Pixal3D base weights       Shared · Installed
Image to 3D               Ready
WorldSculpt               Additional weights required

Runtime contract

Capability identity must remain separate from storage identity.

"MODEL_DIR.name" must not be used as the only way to determine which node is running. That assumption fails as soon as multiple nodes share a directory.

The runtime should receive explicit information equivalent to:

MODEL_ID=<extension-id>/<node-id>
MODEL_NODE_ID=<node-id>
MODEL_DIR=<node-private-directory>
SHARED_MODEL_DIRS={"pixal3d-base":"<resolved-shared-directory>"}

The exact transport may be environment variables or a typed runtime context, but it must work for both direct and subprocess extensions.

Requirements:

- The selected generator remains "<extension-id>/<node-id>".
- "MODEL_DIR" retains its existing node-private meaning for backward compatibility.
- Shared group paths are supplied explicitly by the host.
- Extensions do not construct shared paths from user-controlled strings.
- Updating model storage paths at runtime updates both private and shared directories.

Deletion and uninstall behavior

Shared weights must not be deleted as if they belonged to one capability.

Expected behavior:

- Removing a node's private weights deletes only that node's private directory.
- Removing a shared group requires a separate, explicit action and warns that every dependent node will become unavailable.
- All loaded generators that reference a group are unloaded before that group is removed.
- Uninstalling an extension with “delete model weights” selected removes the extension's model root once, including shared and private data.
- Deleting one capability must never silently remove a group still required by sibling capabilities.
- UI state is refreshed for every affected node after a shared group is downloaded, repaired, cancelled, or removed.

Manifest validation

Reject the extension during validation when:

- a node references an unknown group;
- group IDs are duplicated;
- a group ID or destination is unsafe;
- a group contains invalid or colliding sources;
- checks escape their group root;
- a node attempts to reference a group from another extension;
- shared-weight metadata is declared on a non-model extension in this first phase.

The Python and Electron implementations should apply equivalent normalization and validation rules.

Backward compatibility

- Existing manifests without "weight_groups" keep their current behavior.
- Legacy "hf_repo" and "download_check" fields continue to normalize as node-owned sources.
- Existing node-level "model_sources" remain private to that node.
- Existing model directory layouts remain valid.
- Extensions opt into shared storage explicitly.
- No automatic migration or filesystem deduplication of already downloaded legacy copies is required in the first phase.
- Existing model, workflow, pause/resume/cancel, and uninstall behavior must remain unchanged for manifests that do not use shared groups.

Extension authors can provide migration guidance when changing an existing extension to the shared layout. Safe automatic migration of existing duplicate files can be considered separately.

Acceptance criteria

- [ ] A manifest can declare one or more extension-scoped shared weight groups.
- [ ] Multiple sibling model nodes can reference the same group.
- [ ] A node can combine shared groups with private "model_sources".
- [ ] Installing two nodes that reference the same group downloads that group only once.
- [ ] "Install all" deduplicates shared download work.
- [ ] Readiness is calculated from all shared and private checks required by each node.
- [ ] Runtime node selection does not depend on the shared directory name.
- [ ] Direct and subprocess extensions receive the resolved shared directories.
- [ ] Removing node-private weights preserves shared groups.
- [ ] Removing a shared group warns about and invalidates all dependent nodes.
- [ ] Unsafe, unknown, duplicate, or cross-extension group references are rejected.
- [ ] Existing manifests and node-owned weight directories behave exactly as before.

Suggested regression fixtures

Fully shared checkpoint

Two or more nodes reference one group and have no private sources.

Verify:

- one physical download;
- all nodes become ready;
- concurrent installation cannot start duplicate writes;
- removing the group invalidates all nodes.

This represents the SenseNova use case.

Shared base plus private adapter

Two nodes reference the same base group. One node declares an additional private adapter source.

Verify:

- downloading the base node does not download the adapter;
- the adapter node remains unavailable until its private source is complete;
- installing the adapter node reuses the base;
- deleting the adapter source preserves the base node.

This represents DreamStyle3D and Pixal3D/WorldSculpt.

Legacy manifest

A node uses only "hf_repo", "download_check", or node-level "model_sources".

Verify that download, readiness, runtime paths, deletion, and UI state are unchanged.

Out of scope for phase 1

- Sharing groups between different extensions.
- A global or content-addressed model cache.
- Automatic deduplication or migration of existing files.
- Sharing Python environments, extension source code, or processes.
- Arbitrary local filesystem paths.
- Non-Hugging-Face providers.
- Changes to the model execution API unrelated to resolved weight paths.

A later phase could introduce cross-extension sharing through a separately reviewed global identity, versioning, trust, reference-counting, and garbage-collection contract.

Related work

- #273 / PR #275 introduced validated multi-source downloads owned by one node. Shared groups should reuse that contract rather than add a second downloader.
- #114 includes owner-scoped weights as one part of a much broader bundled-extension proposal. This issue extracts only the single-extension weight-sharing requirement.
- PR #335 experiments with "weight_owner_id", but the shared-weight behavior should be reviewed independently from child-extension installation, text-to-image dispatch, previews, readiness, and VRAM changes.

A direct node-owner alias is not sufficient for this issue because it cannot cleanly represent a shared base plus node-private adapters, and it risks coupling runtime node selection and deletion to a filesystem path.

Checklist

- [x] Existing requests and pull requests were reviewed.
- [x] The first phase is limited to sibling model nodes inside one extension.
- [x] Backward compatibility and filesystem confinement are required.

DrHepa:
Summary

This PR implements extension-scoped shared weight groups for model extensions.

Multiple model nodes within the same extension can now reuse one or more physical weight groups while retaining their own capability identity and optional node-private weights.

This supports both primary use cases described in #343:

- Fully shared checkpoints used by multiple sibling nodes, such as SenseNova.
- A shared base model combined with node-private adapters, such as Pixal3D/WorldSculpt or DreamStyle3D.

Closes #343.

Motivation

Modly currently treats each model node as both a capability and the physical owner of its weights:

<models-dir>/<extension-id>/<node-id>/

This causes every node to download its own copy of the same checkpoint, even when several nodes within an extension use an identical base model.

A simple node-owner alias is insufficient because some nodes need to combine a shared base checkpoint with additional private weights. It would also couple runtime identity and deletion behavior to a physical directory.

This PR separates:

- Capability identity.
- Shared weight ownership.
- Node-private weight ownership.

Manifest contract

Model extensions can declare top-level "weight_groups". Each group owns its own validated "model_sources" plan.

Nodes reference the shared groups they require and may continue declaring private "model_sources".

Example:

{
  "id": "pixal3d",
  "type": "model",
  "weight_groups": [
    {
      "id": "pixal3d-base",
      "model_sources": [
        {
          "id": "pixal3d",
          "provider": "huggingface",
          "repo_id": "TencentARC/Pixal3D",
          "revision": "<pinned-revision>",
          "destination": ".",
          "checks": [
            "pipeline.json"
          ]
        }
      ]
    }
  ],
  "nodes": [
    {
      "id": "generate",
      "weight_groups": [
        "pixal3d-base"
      ]
    },
    {
      "id": "worldsculpt",
      "weight_groups": [
        "pixal3d-base"
      ],
      "model_sources": [
        {
          "id": "worldsculpt-adapter",
          "provider": "huggingface",
          "repo_id": "AlayaLab/WorldSculpt",
          "revision": "<pinned-revision>",
          "destination": "worldsculpt",
          "checks": [
            "worldsculpt/<required-file>"
          ]
        }
      ]
    }
  ]
}

Storage layout

Shared groups are stored independently from node-private weights:

<models-dir>/
  <extension-id>/
    _shared/
      <weight-group-id>/
        ...
    <node-id>/
      ...

Rules enforced by the implementation:

- Shared sources are confined to "_shared/<weight-group-id>".
- Node-private sources remain under "<node-id>".
- Group IDs are local to their extension.
- "_shared" is reserved and cannot be used as a node ID.
- Absolute paths, "..", unsafe IDs, path separators inside IDs and escaping checks are rejected.
- Cross-extension weight-group references are not supported.
- No symlinks or hard links are required.

Download behavior

Installing a node builds an effective download plan containing:

1. Every shared weight group referenced by the node.
2. The node’s private "model_sources".

The download lifecycle now:

- Skips shared groups that are already complete.
- Deduplicates shared work during “Install all”.
- Locks downloads by canonical physical target.
- Prevents concurrent node installations from writing to the same shared group.
- Preserves existing pause, resume, cancellation and repair behavior.
- Keeps a completed shared group valid if a node-private source fails.
- Refreshes all dependent node states when a shared group changes.

Readiness behavior

A node is ready only when:

- Every referenced shared group passes its checks.
- Every node-private source passes its checks.

This allows states such as:

- Pixal3D base installed → the regular Pixal3D node is ready.
- Pixal3D base installed but WorldSculpt adapter missing → WorldSculpt is not ready.
- Installing WorldSculpt reuses the existing Pixal3D base and downloads only its additional weights.
- One completed SenseNova group makes every dependent node without private sources ready.

Runtime contract

Capability identity is now explicitly separated from storage identity.

The runtime no longer depends on "MODEL_DIR.name" to determine which node is running.

Direct and subprocess extensions receive:

MODEL_ID=<extension-id>/<node-id>
MODEL_NODE_ID=<node-id>
MODEL_DIR=<node-private-directory>
SHARED_MODEL_DIRS={"<group-id>":"<resolved-shared-directory>"}

The Python generator context also exposes the resolved "shared_model_dirs" mapping.

"MODEL_DIR" keeps its existing node-private meaning for backward compatibility.

When the configured model storage path changes, both private and shared directories are resolved again from the host configuration.

Deletion and uninstall behavior

Shared groups have an independent lifecycle:

- Removing node-private weights deletes only the node directory.
- Removing a shared group requires a separate explicit action.
- The UI warns that every dependent node will become unavailable.
- Loaded generators referencing the group are unloaded before deletion.
- Removing a shared group refreshes every affected node.
- Deleting one capability never silently removes weights required by sibling capabilities.
- Uninstalling an extension with model-weight deletion enabled removes the extension model root once, including shared and private data.

UI changes

The Models UI and extension drawer now expose:

- Shared weight-group installation state.
- Shared versus node-private weight requirements.
- Nodes requiring additional private weights.
- Explicit shared-group installation and removal actions.
- Warnings listing the nodes affected by shared-group removal.

Manifest validation

The Electron and Python implementations apply equivalent normalization and validation rules.

Validation rejects:

- Duplicate group IDs.
- Unknown group references.
- Unsafe group or node IDs.
- "_shared" as a node ID.
- Unsafe destinations or checks.
- Sources that collide within the same physical root.
- Shared-weight metadata on non-model extensions.
- Attempts to reference groups outside the current extension.

Backward compatibility

Extensions that do not declare "weight_groups" retain their existing behavior and directory layout.

The following remain node-private and fully supported:

- Legacy "hf_repo".
- Legacy "download_check".
- Existing node-level "model_sources".

This PR does not migrate or deduplicate existing downloaded files automatically. Extension authors can provide migration guidance when adopting the new layout.

Test coverage

Regression coverage includes:

- Multiple nodes using one fully shared checkpoint.
- A shared base combined with a private adapter.
- Readiness propagation across dependent nodes.
- Install-all download deduplication.
- Canonical physical-target locking.
- Explicit runtime node identity.
- Direct and subprocess shared-directory propagation.
- Shared-group removal and dependent-node invalidation.
- Preservation of shared weights when private weights are removed.
- Manifest validation and path confinement.
- Compatibility with legacy manifests and node-owned directories.

Validation results

- Python: 118 tests passed.
- Node: 178 tests passed.
- Lint passed.
- Production build passed.

Out of scope

This first phase intentionally does not add:

- Weight sharing between unrelated extensions.
- A global or content-addressed model cache.
- Automatic migration of existing duplicate weights.
- Shared Python environments or extension processes.
- Arbitrary local filesystem paths.
- Non-Hugging-Face providers.

Cross-extension sharing can be designed separately once ownership, versioning, trust, reference counting and garbage-collection semantics have been defined.


iammojogo-sudo:
Is this possible with different venv's though? Each venv has its own dependencies that are weight based for calculations in many  cases. I would LOVE if this would work as it would reduce some memory and space on my laptop. Very curious how this plays out @DrHepa

DrHepa:
Thanks, @iammojogo-sudo! Sharing checkpoint files doesn’t mean sharing a Python environment. This PR only reuses weights within one extension, whose nodes already share a venv. I checked our extensions: SD15/SDXL have matching dependency versions, but others need different Torch/CUDA or Transformers versions, so one shared venv isn’t a safe default. I’ve opened [a separate proposal](https://github.com/lightningpixel/modly/issues/349) to keep venvs isolated while reusing verified package files where possible. That could save disk space—not RAM/VRAM.

iammojogo-sudo:
" so one shared venv isn’t a safe default" exactly! That's why I was wondering. But it is good that "keep venvs isolated while reusing verified package files" because that right there could save a lot of re-downloading python dependencies at the least. Good stuff! If it can be made internal (in the .asar), that would be helpful for extension creators too, so that they don't have to code a shared communication across extensions. Could be a big deal! I love this.

